PR #5, opened 2017-01-16
Implement function key

roadrunner2:
This was tested on a MBP Late 2016, i.e. with touchbar, so haven't tested the F1-F12 keys. This also means the driver works on these new MBP's (modulo a small hiccup) - many thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ The 2015 MacBook seems much more complicated, as the DMA controller isn't built 
 What works:
 -----------
 * Basic Typing
+* FN keys
 * Driver unloading (no more hanging)
 * Basic touchpad functionality (even right click, handled by libinput)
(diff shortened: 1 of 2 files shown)


cb22:
Hey @roadrunner2 

Thanks for the pull request; glad to hear the driver works on the newer MBPs too. I'll give it a test to check the F1-F12 keys on my 12" MB, but otherwise looks good to me.

roadrunner2:
I just realized one thing: we're not actually sending an event for the function-key itself currently. I'm not sure if the input processing will do anything with it, but I noticed ```hid-apple.c``` sends it, so I should probably add that.

roadrunner2:
I just updated the patch to also generate the KEY_FN event.

cb22:
Great, can confirm everything (keyboard side) works well on my MB 12"

Issue #6, opened 2017-01-22
Touchpad not initilized properly on MacBookPro13,3


roadrunner2:
Generally on boot the touchpad is unavailable; one or more '```sudo rmmod applespi && sudo modprobe applespi```' will get it working, though. This seems to indicate some initialization issue. I was able to narrow things down a bit to lines 282 - 285 (sending the ```applespi_init_commands```): if I put a loop around just those to run them 10 times, then I have a functioning touchpad on boot in about 50% of the cases. This obviously isn't a fix or even a usable workaround, of course.

Where did those magic ```applespi_init_commands``` come from? Any suggestions on how to proceed further here? Do we need to somehow trace the driver in MacOS?

cb22:
The magic command to perform the modeswitch came from monitoring what the Windows driver does and basically sending the same command. It actually does a whole host of other commands (such as getting the name and vendor of the device), but that seemed to be the one that switched it in to multitouch mode. 

[Here's](https://github.com/cb22/macbook12-spi-driver/issues/3#issuecomment-248405522) how I got the Windows driver to dump the data it was sending over the bus. Each command does have a response associated with it. 

It might work simply sending the modeswitch command repeatedly until receiving back the correct response. However, the Windows driver seems to not have to do such things, which leads me to believe there's something else (timing? additional reset commands?) at play.

Unfortunately, I don't have any deeper insight in to the protocol format at this point.

Dunedan:
I also had lots of troubles with touchpad initialization. While I noticed that altering the touchpad initalization lead to no improvement for me I found something else what fixed the problem for me every time so far. That's simply:
```
rmmod spi_pxa2xx_platform && modprobe spi_pxa2xx_platform
```

Dunedan:
The solution mentioned above is also only working the same percentage of time as reloading the `applespi` driver itself. :disappointed: 

Another thing I noticed is that even while the touchpad is in that not properly initialized state, touches are somehow registered. They don't act as clicks, but at least @roadrunner2 `appletb` driver is capable of waking up the TouchBar again on such a press. I don't know if that's already known, but I thought it's worth mentioning it.

cb22:
@Dunedan Yeah, so what happens is the touchpad can operate in two modes; one providing only basic functionality, and other giving full multitouch functionality.

[...]

Issue #9, opened 2017-01-25
2015 Macbook testing


theishshah:
It seems that the driver is primarily non-functional on the 2015model. The trackpad is essentially broken and (possibly random, need more testing) very few keystrokes are registering.

cb22:
@theishshah interesting; so it was registering some keystrokes? Last I heard about the 2015 model the SPI driver wouldn't bind properly

theishshah:
Yeah I'm not sure it seems very sporadic. It is possible that it was interpreting input from an external usb device or something, i'm not sure

Maximus-:
Under a stock linux kernel (4.11.0-rc6) with Intel LPSS modules disabled, the keyboard works again with this module on the 2015 model. However, there are a few issues. Typing is kind of slow/laggy, which could be due to systemd, but it seems unlikely. Also there are issues when typing too fast, which may be due to rollover previously mentioned. Also, trackpad doesn't work at all. It appears the data read related to trackpad is correct, but multitouch isn't properly enabled, as there's only one finger's worth of data.

Maximus-:
Update: the typing lag was due to the clock speed being set low. Bumped it up from 400kHZ to the 8MHz and now typing is nice and smooth. Key-rollover issues still occur, of course. I've got the trackpad working with the rel_x and rel_y parameters passed from the device, however I'm getting no multitouch data.

cb22:
@Maximus- interesting you say about bumping up the speed; the multitouch initialization might have to do with the issue mentioned by @roadrunner2 [here](https://github.com/cb22/macbook12-spi-driver/issues/6#issuecomment-294447181). 

Alternatively, the initialization packet might be difference on the 2015 model. That would need to be decoded from Windows and added to the driver.

Maximus-:
I'm trying to find those packets in the OS X kexts, and I've checked AppleHSSPIHIDDriver, AppleHSSPISupport, AppleIntelLpssDmac, and AppleIntelLpssGspi; I can't find any trace of them, but they may just be stored a bit differently in the binaries. Downloading the windows drivers now. 

Also some things I'm not entirely clear on:
What's the actual difference between the two machines (2015, and 2016 models)?
Why does this work when all Intel LPSS-related kernel modules are disabled? 
Furthermore, should I care about having LPSS? Battery life seems fine under linux so far.

roadrunner2:
[...]

Issue #3, opened 2016-09-16
Caps Lock LED doesn't work


cgutman:
We probably need to capture the SPI transfers that happen on Windows and macOS to turn the LED on and off to make this work.

cb22:
I captured the packet needed to do this a while back, let me dig them up. If you'd like to experiment with sniffing the SPI data yourself, here's how I did it in Windows 10:

> It's possible to get Windows to dump everything it sends on what it calls SPB - basically SPI & I2C. Open the Event Viewer and enable View -> Show Analytic and Debug Logs. Then head to Application and Services Logs -> Microsoft -> Windows -> SPB-ClassExtension. Right click on Analytic and enable it - all events will be logged. Disable it when done, and look for the 1023 events, and view the details tab - they'll give you the full buffer contents. Other events will let you know things like the direction of the transfer, etc.

There are still quite a few packets with unknown contents. Definitely a cool one to reverse would be the "force click" that happens; but to the best of my knowledge that only happens under macOS.

roadrunner2:
Can you dig up those traces? I'd love to add this functionality (and submit a pull request).

cb22:
@roadrunner2 I don't think I have them anymore, but I'll do a capture from Windows. Sorry, I haven't had much time to give to this driver as of late

roadrunner2:
I got some traces from @tudorbarascu over on issue #6, and am making progress on this - the basics are working, but still have some reliability issues. Will send a pull request when it's ready.

PR #14, opened 2017-04-23
Init fixes and caps lock led support

roadrunner2:
Apologies for sending this in one pull request, but I'm too lazy to try and tear things apart. The two big features where are the fixes to make trackpad initialization reliable, and to toggle the caps-locks led. Besides that there are some cleanups/re-factorings that I hope are acceptable.

This should fix issues #6 and #3, and is also related to #9.

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -336,9 +336,9 @@ applespi_check_write_status(struct applespi_data *applespi, int sts)
 	u8 sts_ok[] = { 0xac, 0x27, 0x68, 0xd5 };
 
 	if (sts < 0)
-		pr_warn("Error writing to device: %d", sts);
+		pr_warn("Error writing to device: %d\n", sts);
 	else if (memcmp(applespi->tx_status, sts_ok, APPLESPI_STATUS_SIZE) != 0)
-		pr_warn("Error writing to device: %x %x %x %x",
+		pr_warn("Error writing to device: %x %x %x %x\n",
 			applespi->tx_status[0], applespi->tx_status[1],
 			applespi->tx_status[2], applespi->tx_status[3]);
 }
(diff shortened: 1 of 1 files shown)


cb22:
Great! This looks good to me. I'm without my laptop for a little bit (screen issue getting fixed), so I can't test it out myself - but if you're satisfied, then I'm happy to merge it in.

roadrunner2:
@cb22 I'm reasonably happy with the changes, but it would nevertheless be good if somebody could test them on a MacBook (8 or 9), in particular the caps-lock led functionality (the traces I based it off are from a MacBookPro13,* - it's quite likely they're the same on the MacBook's, but it's just not been tested there).

Dunedan:
Really awesome work. Hit the caps lock key multiple times today just to see the fancy LED light up. :smile:

Looking at the diff the SPI init and the caps lock LED commands look pretty similar. Any idea if there are additional commands available?

Dunedan:
One probably pretty minor issue I noticed: When reloading the `applespi` driver the `Xorg.0.log` gets filled with the following message for a few milliseconds:
```
[ 38290.618] (EE) Apple SPI Touchpad: Read error 19
```

roadrunner2:
@cb22 Everything has been confirmed working on a MacBook8 (see #9), so from my side I'm happy now.

roadrunner2:
@Dunedan Yes, the structure of the commands are quite similar - I've been able to deduce a few more things about the structures of the messages in general (e.g. I understand the 8-byte header now, and know how to detect and deal with continued messages, such as occur when more than 6 fingers are pressed), and plan to submit a pull to make use of that and do a little less black-magic bit banging. And in fact there are some other commands in the traces @tudorbarascu sent, but I haven't tried to see what they do.

As to the error: hmm, errno 19 is ```ENODEV```, and those are Xorg messages, so not sure there's anything in the driver that can be done about this.

PR #17, opened 2017-04-24
add option iso_layout

### LINKED ISSUE #12: Missing handling for ISO-layout keyboard
I was wondering since a while why two keys on the keyboard of my MacBookPro 13,2 weren't working as expected. I finally figured it out! It's because `applespi` currently doesn't handle keyboards with ISO-layout differently than with ANSI-layout. As I've a german keyboard that's the culprit.

If you look at `hid_apple` for USB-connected Apple keyboards, you'll see the special handling they do for ISO-layout keyboards:
https://github.com/torvalds/linux/blob/master/drivers/hid/hid-apple.c#L153-L157

In addition `KEY_100ND` isn't detected, as it uses the scancode `100`, which currently isn't part of `applespi_scancodes`. I was able to get it properly working with the following patch:
```patch
diff --git a/applespi.c b/applespi.c
index 4356f84..3bf40ea 100644
--- a/applespi.c
+++ b/applespi.c
@@ -129,11 +129,12 @@ static const unsigned char applespi_scancodes[] = {
        KEY_1, KEY_2, KEY_3, KEY_4, KEY_5, KEY_6, KEY_7, KEY_8, KEY_9, KEY_0,
        KEY_ENTER, KEY_ESC, KEY_BACKSPACE, KEY_TAB, KEY_SPACE, KEY_MINUS,
        KEY_EQUAL, KEY_LEFTBRACE, KEY_RIGHTBRACE, KEY_BACKSLASH, 0,
-       KEY_SEMICOLON, KEY_APOSTROPHE, KEY_GRAVE, KEY_COMMA, KEY_DOT, KEY_SLASH,
+       KEY_SEMICOLON, KEY_APOSTROPHE, KEY_102ND, KEY_COMMA, KEY_DOT, KEY_SLASH,
        KEY_CAPSLOCK,
        KEY_F1, KEY_F2, KEY_F3, KEY_F4, KEY_F5, KEY_F6, KEY_F7, KEY_F8, KEY_F9,
        KEY_F10, KEY_F11, KEY_F12, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        KEY_RIGHT, KEY_LEFT, KEY_DOWN, KEY_UP,
+       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, KEY_GRAVE,
 };
 
 static const unsigned char applespi_controlcodes[] = {
```

This patch wouldn't work for ANSI-keyboards, as they don't have the additional key with scancode 100, therefore we need some kind of key mapping like `hid_apple` does.

christophgysin:
fixes #12

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -45,6 +45,10 @@
 #define MAX_FINGERS		6
 #define MAX_FINGER_ORIENTATION	16384
 
+static unsigned int iso_layout = 1;
+module_param(iso_layout, uint, 0644);
+MODULE_PARM_DESC(iso_layout, "Enable/Disable hardcoded ISO-layout of the keyboard. "
+		"(0 = disabled, [1] = enabled)");
 
 struct keyboard_protocol {
 	u16		packet_type;
@@ -177,6 +181,7 @@ static const unsigned char applespi_scancodes[] = {
 	KEY_F1, KEY_F2, KEY_F3, KEY_F4, KEY_F5, KEY_F6, KEY_F7, KEY_F8, KEY_F9,
 	KEY_F10, KEY_F11, KEY_F12, 0, 0, 0, 0, 0, 0, 0, 0, 0,
 	KEY_RIGHT, KEY_LEFT, KEY_DOWN, KEY_UP,
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, KEY_102ND,
 };
 
 static const unsigned char applespi_controlcodes[] = {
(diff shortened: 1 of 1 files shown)


cb22:
Thanks for the PR @christophgysin! Could you perhaps fix the conflicts with the recently merged PR, then I'd be happy to merge it in.

christophgysin:
Conflicts resolved.

Dunedan:
I tried out this PR and while it works in general the option seems to work the opposite it should. If I use the default or explicitly set `iso_layout=1` I get the keys for the non-iso layout. Setting it to `iso_layout=0` makes it working as it should on a iso layout keyboard.

christophgysin:
I'm not sure which layout I have, but it seemed to behave identically to the option on my external apple keyboard using hid-apple.ko

I will double check tomorrow. Inverting the logic would be trivial.

Dunedan:
I think I got what's causing the bug. With `iso_layout=1` you switch the keys, but you already switched them once directly in your PR, when you replaced `KEY_GRAVE` with `KEY_102ND`. Simply switch the order of keys in `applespi_scancodes` back again and it should work as expected.

If you want to know which layout you have you can simply refer to: https://deskthority.net/wiki/ANSI_vs_ISO

christophgysin:
I undid switching the keys in applespi_scancodes and pushed a new commit.

Dunedan:
Works as expected now. :+1:

christophgysin:
@cb22 Is there anything else I can do to get this merged?

cb22:
Just on holiday at the moment without reliable internet - as soon as I'm
back I'll merge it in 👍

roadrunner2:
I just updated to this, and I don't think the current situation is reasonable: by default it now assumes an iso-layout, which I think is backwards (I'm fairly certain non-iso layouts are more common). Now, the default value of 1 (enabled) is the same as the stock `hid-apple` driver has, **but** that driver also detects whether the keyboard is actually an iso one or not, i.e. does the right thing automatically - that is not the case here currently.

I therefore think that either the default needs to be changed (0, disabled), or we need figure out how to detect which keyboard is present.

Regarding the latter, I see two potential candidates: some dmi information, or information returned for some spi keyboard command. The only dmi item I think that might encode this information is the `board_name` or `chassis_version` (these contain the same value on my system, a MacBookPro13,3 with regular keyboard: `Mac-A5C67F76ED83108C`). As to SPI keyboard info commands, I don't know if these even exist, but I think it's likely - we'd need traces from early system boot to try and find candidates.

PR #19, opened 2017-04-27
add option fnmode

christophgysin:


christophgysin:
This also includes the commit that adds the option iso_layout. Ideally, that PR is merged first.

Dunedan:
I really like your approach of bringing `applespi` to feature-parity with `hid-apple`, but this PR somewhat collides with the `fn_mode` option from the `appletb` driver (https://github.com/roadrunner2/macbook12-spi-driver/blob/touchbar-driver-hid-driver/appletb.c#L60-L64). I haven't noticed unexpected behavior using `applespi` with this patch in combination with `appletb`, but maybe you can sync with @roadrunner2 before going on, to have a common solution.
I'd prefer naming and behavior identical to `hid_apple` so users who switch from an older to a newer Mac only have to replace the kernel module and can keep its options as they are.

christophgysin:
@Dunedan appletb is a separate module with its own set of module parameters, so there shouldn't be any collision. Or am I missing something?

Dunedan:
Yes but my thought was that it'd be nice if they would share the same parameters for configuring the same functionality. I mean the only difference is that one driver is for physical function keys, while the other one is for virtual ones.

christophgysin:
@roadrunner2 Since hid-apple has been around for a while, I'd suggest adopting its module parameter names and values. What do you think?

roadrunner2:
I guess I'll rename the module parameter to 'fnmode' then, even though the current name ('default_fn_mode') is more accurate.

christophgysin:
Any chance getting this merged?

cb22:
@christophgysin sorry about the delay, thought I had merged this in already

christophgysin:
Thanks!

PR #28, opened 2017-05-28
applespi: Change the default for iso-layout to disabled.

roadrunner2:
Unlike the hid-apple driver, we don't have any idea whether the current
keyboard is using the iso-layout or not. This means that the iso_layout
has subtly different semantics here than on hid-apple (from which this
was copied): rather than just allowing for disabling of the iso-layout
on iso keyboards, the flag here determines which layout is used on all
keyboards. Therefore the flag's default should reflect the most common
keyboard type, which is not iso.

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -52,7 +52,7 @@ module_param(fnmode, uint, 0644);
 MODULE_PARM_DESC(fnmode, "Mode of fn key on Apple keyboards (0 = disabled, "
 		"[1] = fkeyslast, 2 = fkeysfirst)");
 
-static unsigned int iso_layout = 1;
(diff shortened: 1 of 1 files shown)


roadrunner2:
Forgot to mention: see also my comments in #17 .

tudorbarascu:
@roadrunner2 just tried it and I can't seem to get it to work.

I'm getting the ``applespi: No spi-master device found for bus-number 2 - waiting for it to be registered`` when doing ``dmesg``

cb22:
I agree - I missed the fact that `iso-layout` got set to the default in the earlier PR, so thanks for this.

PR #31, opened 2017-06-12
applespi: Add keyboard-backlight support.

roadrunner2:
I assume you already started adding this functionality to the driver, but I got impatient and too excited about having the backlight working after you posted your traces in Dunedan/mbp-2016-linux#16 that I went ahead and created this. Feel free to ignore, trash, plunder, or otherwise do what you like with this.

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -31,6 +31,7 @@
 #include <linux/version.h>
 #include <linux/workqueue.h>
 #include <linux/notifier.h>
+#include <linux/leds.h>
 
 #include <linux/input.h>
 #include <linux/input/mt.h>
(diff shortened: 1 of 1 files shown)


Dunedan:
Works fine. :+1:

cb22:
@roadrunner2 this is great! Confirmed working on my Macbook9,1, so I'm going to go ahead and merge it. 

Really oddly, it seems like the SPI issues I was having previously have vanished. No idea what was going on there.

roadrunner2:
@cb22 Interesting about the SPI issues vanishing. I suppose that's both good and bad news: good that it's working for you now, but bad in that it's leaving me worried we still have some sporadic issue here and no way to debug it. Oh well, will have to wait and see if anybody else complains of the same problem.

Thanks for merging this and all the other pull requests!

PR #40, opened 2017-07-24
Touchpad dimensions updates

roadrunner2:
The 3 changes here are actually independent, but are all related to touchpad dimensions, so I thought it might be easier if I sent them as one pull request.

The logging of touchpad dimesions is a bit more complicated using dynamic-debug instead of simple #ifdef's, but I think the advantage of being able to give folks a simple echo command to run to turn it on instead of asking them recompile outweighs the uglyness - this should be especially true once the module is mainlined and folks get it as part of their distro's kernel.

I'm still not happy about the MB touchpad dimesions. As I mention in the commit message, the fact that the old values and those reported for the MB10,1 have noticeably different aspect ratios doesn't sit well (see also [my comments](https://github.com/cb22/macbook12-spi-driver/issues/9#issuecomment-312547951) in #9). It could also be the dimensions for the older MB's are wrong, though I assume you acquired them correctly (but maybe a quick double check using the logging added in this pull request might not hurt).

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -303,6 +303,30 @@ static const struct dmi_system_id applespi_touchpad_infos[] = {
 		.driver_data = &applespi_macbookpro133_info,
 	},
 	{
+		.ident = "Apple MacBookPro14,1",
+		.matches = {
+			DMI_MATCH(DMI_SYS_VENDOR, "Apple Inc."),
+			DMI_MATCH(DMI_PRODUCT_NAME, "MacBookPro14,1")
+		},
+		.driver_data = &applespi_macbookpro131_info,
+	},
+	{
+		.ident = "Apple MacBookPro14,2",
+		.matches = {
+			DMI_MATCH(DMI_SYS_VENDOR, "Apple Inc."),
+			DMI_MATCH(DMI_PRODUCT_NAME, "MacBookPro14,2")
+		},
+		.driver_data = &applespi_macbookpro131_info,	// same touchpad
+	},
+	{
+		.ident = "Apple MacBookPro14,3",
+		.matches = {
+			DMI_MATCH(DMI_SYS_VENDOR, "Apple Inc."),
+			DMI_MATCH(DMI_PRODUCT_NAME, "MacBookPro14,3")
+		},
+		.driver_data = &applespi_macbookpro133_info,
+	},
+	{
 		.ident = "Apple Generic MacBook(Pro)",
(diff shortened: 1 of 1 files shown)


cb22:
> The logging of touchpad dimesions is a bit more complicated using dynamic-debug instead of simple #ifdef's, but I think the advantage of being able to give folks a simple echo command to run to turn it on instead of asking them recompile outweighs the uglyness - this should be especially true once the module is mainlined and folks get it as part of their distro's kernel.

Agreed.

> I'm still not happy about the MB touchpad dimesions. As I mention in the commit message, the fact that the old values and those reported for the MB10,1 have noticeably different aspect ratios doesn't sit well (see also my comments in #9). It could also be the dimensions for the older MB's are wrong, though I assume you acquired them correctly (but maybe a quick double check using the logging added in this pull request might not hurt).

Actually, I'm not entirely sure they are correct myself. I might have just lifted them straight from the BCM5974 driver in my initial excitement at getting something working. 

I'll run this branch on my MB9,1 and report back shortly 👍

cb22:
So I just ran it on my 9,1 which reports

```
 -5087 5579 -182 6089
```

Compared to the current values of

```
-4828, 5345, -203, 6803
```

So those current values in the code are definitely off. The ones I get are basically the same as what you've got for the 10,1

spectacle m25314

roadrunner2:
Ah, this is good! I would then suggest we pick the largest dimensions found (it can be tricky to trigger those "pixels" at the very edge) and use them for all MB's, i.e. the measurements you got.

Also, can you confirm the physical dimensions of your trackpad? The were reported as 70mm x 112mm.

roadrunner2:
Btw., I just realized I managed to mess up the last patch in this series. Anyway, I reworked and repushed the last patch so it now uses your above reported values for all MacBook's.

cb22:
I can confirm the dimensions as 70mm x 112mm (measured using some calipers) 👍

roadrunner2:
Thanks. So that makes the x resolution 95 and the y resolution 90 - I still think it's odd for these two values to be so different here when they are only 1 or 2 points different for all other Mac's, but I guess this must just be something with this touchpad. Anyway, I've updated my [evdev hwdb override](https://gist.github.com/roadrunner2/1289542a748d9a104e7baec6a92f9cd7#file-61-evdev-local-hwdb) appropriately now, and will submit it upstream soon.

PR #41, opened 2017-07-24
applespi: Fix reading of touchpad init response.

roadrunner2:
This wasn't actually causing any issues AFAICT, since we currently don't look at the response, but nevertheless I figure it doesn't hurt.


cb22:
Good catch. We probably should start looking at the response. I believe you said you made some good headway in decoding the packet format properly?

PR #42, opened 2017-07-24
applespi: Convert #ifdef'd debug logs to normal debug logs.

roadrunner2:
I'm submitting this mainly for discussion rather than for actual merging. The biggest problem I see with this as it stands is that it's hard to enable debugging of just some packets, and enabling all results in a _huge_ amount of logging in the kernel log, swamping pretty much everything else (mostly due to the touchpad events). E.g. while it's possible to enable logging of just the touchpad init messages, that's only because they are currently only ones using the `applespi_sync_*` functions, i.e. if something else starts using them in the future then that new stuff would get logged to when enabling logging for these functions. And enabling logging of received events does so for both keyboard and touchpad events. Also, as I've discovered recently, it doesn't seem possible to enable logging of just some functions via the kernel command line (the docs indicate it should be, but after debugging the dynamic-debuging code in the kernel I've come to the conclusion that just doesn't seem to be the case).

I think ideally we'd want a way to be able to (dynamically) enable/disable logging of messages based on type: init, keyboard events, touchpad events, backlight commands, etc. But AFAICT this can't be done cleanly with the dynamic-debug framework alone. So, maybe we need our own `debug` module parameter to control logging of messages.

Comments, ideas, thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,15 @@ Touchpad:
 ---------
 The touchpad protocol is the same as the bcm5974 driver. Perhaps there is a nice way of utilizing it? For now, bits of code have just been copy and pasted.
 
+Debugging:
+----------
+The `debug` module parameter can be used to turn debugging output on (and off) dynamically, and can be set in all the usual ways (e.g. via kernel command-line (`applespi.debug=0x1`), via sysfs (`echo 0x10000 | sudo tee /sys/module/applespi/parameters/debug`), etc.).
+
+Some useful values are (since the value is a bitmask, these can be combined):
+* 0x10000 - determine touchpad values range
+* 0x1     - turn on logging of touchpad initialization packets
+* 0x6     - turn on logging of backlight and caps-lock-led packets
+
 Some useful threads:
 --------------------
 * https://bugzilla.kernel.org/show_bug.cgi?id=108331
--- a/applespi.c
+++ b/applespi.c
@@ -45,7 +45,6 @@
 
 #define PACKET_KEYBOARD         288
 #define PACKET_TOUCHPAD         544
-#define PACKET_NOTHING          53312
 
 #define MAX_ROLLOVER 		6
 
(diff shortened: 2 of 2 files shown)


cb22:
@roadrunner2 hmm. Perhaps it might be useful before things are very well tested to allow full packet tracing to be turned on at runtime? My thinking is that if weird things start happening to the trackpad / keyboard, being able to SSH in and enable debugging on the fly (without having to reload the module) might prove to be useful.

That being said, I'm not sure how common (if it all) said issues are. Judging from your comments in #6 it seems like you're getting to the bottom of strange timing issues. In that case, I think just having a `debug` parameter for when the driver flat out refuses to work is a fair way to go.

roadrunner2:
@cb22 Yes, absolutely: it must be controllable at runtime. My thinking is to create a "debug" module parameter that is a bitmap to control various bits of debug logging (init, caps-lock control, backlight control, keyboard event, touchpad event, unknown event); and by givng this param mode 644 it could be changed dynamically (by writing to `/sys/modules/applespi/parameters/debug`), just like the current `iso_layout` and `fnmode` params can. Compared to dyndbg this gives a more stable "interface" for enabling bits (not affected by function renaming or refactoring) and allows arbitrary debugging to be enabled at boot time (for some reason the dyndbg `func` and `line` match specs don't work when provided on the kernel command line), while still allowing full runtime control.

I'm also thinking that the touchpad dimensions logging should be changed to use this `debug` param, for consistency.

Thoughts?

roadrunner2:
Here is a new version of this pull requests. As discussed above, instead of using the dynamic-debug facilities debugging is now controlled via a simple module parameter 'debug'. I added some notes about it to the README too, but you can now enable various debug logging simply with something like
```
echo 0x7 | sudo tee /sys/module/applespi/parameters/debug
```
(this would start logging all "commands", including touchpad init, caps-lock led, and backlight).

Logging of touchpad dimensions is now also enabled via this parameter.

I have already found this to be very useful in debugging the delay issues.

cb22:
@roadrunner2 great! This looks good to me, so I'm going to go ahead and merge it in.

roadrunner2:
Thanks.

Issue #44, opened 2017-08-05
Driver should autosense ISO layout


l1k:
Commit a7a73f0dbd485318b0f71405cf350106a6f826c7 introduced a command line parameter whether to assume an ISO keyboard layout.

For USB keyboards, the mainline kernel gained the ability to *autosense* an ISO layout with commit torvalds/linux@bd77a0f08ec57f7b805dfbaa64b36329dfa005d6 by looking at the `bCountryCode` field in the HID descriptor: 0x13 means ISO layout and e.g. 0x0d means German (de_DE) layout.

It seems likely that Apple used the same values and somehow communicates the country code either upon initialization or with every packet. The first step would thus be to look for 0x13 in data coming from the keyboard controller on a MacBook with ISO keyboard, then check if a German keyboard has 0x0d in the same byte position.

See also the <a href="https://patchwork.kernel.org/patch/9762845/">mailing list discussion</a> on the patch for USB keyboards.

jonas2515:
Just a little heads up here: I'm currently looking into why my Apple aluminium USB keyboard with german layout has those swapped keys and I've been digging through kernel history the whole day.

Please do *not* include a parameter that is necessary for unbreaking keyboards and leave them broken by default. That was done with the `hid_apple` `iso_layout` parameter and now we have a total mess where some keyboards work and some haven't been working for years, but nobody actually bothers to fix it because there's a parameter that can be set as a workaround.

PR #46, opened 2017-09-17
Updates for 4.14

roadrunner2:
This contains the work to exclude the DSDT hacks on kernels 4.14+, as promised in https://github.com/cb22/macbook12-spi-driver/pull/29#issuecomment-327672918 - tested on 4.14-rc1. Also included are some README updates and a small #ifdef fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
-Very simple, work in progress input driver for the SPI keyboard / trackpad found on 12" MacBooks. 
+Very simple, work in progress input driver for the SPI keyboard / trackpad found on 12" MacBooks (2015 and later) and newer MacBook Pros (late 2016 and later).
 
 Using it:
 ---------
-To get this driver to work on a 2016 12" MacBook or a 2016 MacBook Pro, you'll need to boot the kernel with `intremap=nosid` if you're running a kernel before 4.11. Make sure you don't have `noapic` in your kernel options.
+If you're on a MacBook8,1 (2015) you may need the `irqpoll` kernel parameter. To get this driver to work on all other MacBooks and MacBook Pros, you'll need to boot the kernel with `intremap=nosid` if you're running a kernel before 4.11. In all cases make sure you don't have `noapic` in your kernel options.
 
-Additionally, you need to make sure the `spi_pxa2xx_platform` and `intel_lpss_pci` modules are loaded. This should result in the intel-lpss driver attaching itself to the SPI controller. 
+On the 2015 MacBook you need to (re)compile your kernel with `CONFIG_X86_INTEL_LPSS=n` if running a kernel before 4.14. On all kernels you need ensure the `spi_pxa2xx_platform` module is loaded (if you don't have that module, rebuild your kernel with `CONFIG_SPI_PXA2XX=m`).
 
-On the 2015 MacBook you currently need to (re)compile your kernel with `CONFIG_X86_INTEL_LPSS=n`.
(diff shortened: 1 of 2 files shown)


cb22:
@roadrunner2 great, this looks good to me. Unfortunately I cannot test at the moment, since I'm away and I've messed up my Linux installation (again). 

I'm happy to merge it in if you are. 

I guess at this point, we could probably start looking at doing some cleanup of the driver and getting it ready to send it upstream, now that it can attach without any hacks? (or not?)

roadrunner2:
@cb22 On the one hand these changes are quite safe - other than a little code shuffling and conditional compilation, nothing changes. So yes, I'm quite comfortable with them. On the other hand there's no big hurry to get these in: thing will work find on 4.14 even without these changes (it'll just keep using my hack instead of the new core features). So I'm fine either way, merging now or later.

Regarding cleanup, agreed: I've had that pretty much next on my list. Besides cleanup there is one piece of functionality mising: the code currently assumes the touchpad data always arrives in single 256-byte packet, but when more than 6 fingers are pressed (and the palm sometimes triggers that too) it can be split among 2 256-byte packets. As I head mentioned elsewhere I managed to figure out a few more of the header fields, which provide the info to recognize continuation packets, so I need to make some changes to handle that. After that there's some small stuff like #44 , though I think we could consider upstreaming even without that.

Dunedan:
I can confirm the driver continues to work fine with this PR under Linux 4.14rc1. :+1:

l1k:
Regarding upstreaming, a few things that caught my eye while briefly looking over the code:

- `applespi_setup_spi_message()` is called on every read and write even though the message seems to be identical every time? Call this once for `rd_m` and `wr_m` during `->probe`.
- Can the code be rearranged to get rid of the forward declaration for `applespi_send_cmd_msg()`?
- There are some `//` comments but kernel coding style uses `/* */`
- In how far can the code duplicated from the BCM5974 driver be reduced?

roadrunner2:
@1lk Thanks for the feedback. My responses are:
* calling `applespi_setup_spi_message()` once: I don't think that's safe, as the message is modified by the spi driver (at the very least the status and some length fields); while it may be possible to trace the code and determine exactly what is modified, I think that would be brittle and unsafe.
* removing forward declaration of `applespi_send_cmd_msg()`: no, that can't be removed, or least not without replacing it with a different forward declaration. The reason is that there is a depedency loop: that function passes a callback, `applespi_async_write_complete()`, which calls `applespi_cmd_msg_complete()`, which in turn calls `applespi_send_cmd_msg()`. So one of these needs a forward declaration.
* comments: agreed, fixed, together with some other formatting issues.
* removing duplicated code: not sure there's much duplicated code left, but would have to investigate more closely.

l1k:
<blockquote>calling applespi_setup_spi_message() once: I don't think that's safe, as the message is modified by the spi driver (at the very least the status and some length fields); while it may be possible to trace the code and determine exactly what is modified, I think that would be brittle and unsafe.</blockquote>

A lot of IIO drivers do that, see e.g. <a href="https://git.kernel.org/pub/scm/linux/kernel/git/jic23/iio.git/tree/drivers/iio/adc/mcp320x.c?h=testing">mcp320x.c</a>: `struct mcp320x` contains the `spi_message`, two `spi_transfer`s and the `tx_buf` and `rx_buf`. The `->probe` hook initializes the `spi_transfer`s with the buffers, then initializes the `spi_message` with the `spi_transfer`s. The `spi_message` is then used again and again and again by `mcp320x_adc_conversion()`.

roadrunner2:
I see - thanks for the pointer. I've made this change now (see #48).

PR #47, opened 2017-10-08
applespi: Ensure initrd is rebuilt after installing this driver.

roadrunner2:
When installing a new kernel the initrd tends to get built before the dkms hooks are run, resulting in an initrd that is missing this driver - with this the initrd then gets rebuilt after building the driver for the new kernel.


omnibrain:
I can confirm this, my keyboard stopped working after installing a new kernel even though the dkms hooks ran.

PR #48, opened 2017-10-14
Formatting cleanup

roadrunner2:
Just some basic cleanups. No functional change.

### DIFF
--- a/applespi.c
+++ b/applespi.c
@@ -88,12 +88,12 @@
 static unsigned int fnmode = 1;
 module_param(fnmode, uint, 0644);
 MODULE_PARM_DESC(fnmode, "Mode of fn key on Apple keyboards (0 = disabled, "
-		"[1] = fkeyslast, 2 = fkeysfirst)");
+		 "[1] = fkeyslast, 2 = fkeysfirst)");
 
 static unsigned int iso_layout = 0;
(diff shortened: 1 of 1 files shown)


roadrunner2:
Added another commit on top here in response to @l1k's comments. It's not a formatting cleanup, but it just refactors the code a bit, so still no functional change.

l1k:
Just ran this through `scripts/checkpatch.pl --strict -f`, it only finds this nit:
```
CHECK: 'seperate' may be misspelled - perhaps 'separate'?
#1289: 
	/* Now, set up the touchpad as a seperate input device */
```

l1k:
To mainline, put it somewhere in drivers/input/ (keyboard? misc?), add Kconfig and Makefile entries, add copyright + license header to the file, write a commit message which explains a bit which machines this is for and how it works, add Signed-off-by for all contributors, send it to the list and maintainers, cc everyone who might be relevant.

git only allows a single author unfortunately, that's probably either of Federico or Ronald.

The patch might have to go through the merry-go-round a couple of times, depends on the maintainers, some are lax and some are picky. I've no past experience with the input maintainers.

Once the driver gets merged, it will be a bit more cumbersome to make changes as everything goes through the regular review process on mailing lists, instead of informal pull requests on GitHub.

Let me know if you need help with any of these steps.

PR #54, opened 2017-12-09
Second round of cleanups.

roadrunner2:
This is a second set of code cleanups - the largest part is the ones flagged by checkpatch (so hopefully that will avoid style complaints when we attempt to upstream this 😉 ).

After this I have some small fixes and enhancements lined up, plus finally the refinement of the packet/message data structures (including support multi-packet messages, and with that for more than 6 fingers).


applespi.c
```diff
@@ -1,3 +1,13 @@
+/*
+ * MacBook (Pro) SPI keyboard and touchpad driver
+ *
+ * Copyright (c) 2015-2017 Federico Lorenzi
```
Dunedan: @roadrunner2: Given the work you put into that driver, you should add yourself there as author as well. You truly deserve it!

roadrunner2: Thanks, but it's fine. Federico did the original work, without which I wouldn't have had anything to contribute to.

l1k:
So this is almost too minor to mention, but just to avoid objections once it gets posted to the mailing list, in 7e51eac, for multi-line comments, kernel convention is to leave the first line empty and for single-line comments to not use caps, e.g.:
```
	/*
         * The data is stored in pairs of items, first a string containing
	 * the name of the item, followed by an 8-byte buffer containing the
	 * value in little-endian.
	 */

	/* check released keys */
```

The GPL license text at the top can be replaced with:

```
	SPDX-License-Identifier: GPL-2.0
```

Everything else LGTM. Thanks!

applespi.c
cb22: I definitely agree with @Dunedan here, you've done all the heavy lifting into getting the reverse engineered proof of concept I had into shape!

cb22:
Looks good to me, thanks @roadrunner2!

roadrunner2:
@l1k Thanks for the feedback - don't worry about nit-picking, that's fine. Fixed the comments now too, and the license indicator.

@cb22 Pushed a new version of the branch: the changes are just the comment fixes (a new commit), as well as adjusting the copyright header.

PR #66, opened 2018-05-24
Remove workarounds for Linux <4.14

Dunedan:
As a preparation for upstreaming the driver, this commit removes the
workarounds needed to use it with Linux <4.14, as such workarounds
probably won't be accepted upstream.

I'm not entirely sure if this PR should be merged right now or closer
before upstreaming, but at least here it is already.


roadrunner2:
The intent by all means was/is to remove the workarounds in the version submitted upstream. But I don't think we should be removing them here, since folks with older kernels may well be trying use this.

Dunedan:
I believe this driver should get mainlined as soon as possible. While it's not perfect yet (and maybe it'll never be), it works flawlessly for day to day use. Not having it included upstream is a major hurdle for new users to get Linux running on the latest MacBooks and MacBook Pros, as their input devices simply don't work out of the box.

When submitting it for inclusion in the mainline kernel I guess we all agree that it's worth keeping this repository for further development. When doing so I believe it'll be way easier to maintain only a single version of the driver (let's call it _mainline-next_) in addition to the mainline version, instead of two (_mainline-next_ and _compatible-with-pre-4.14_). As resources are scarce, they should be focused on what matters and that's from my point of view having the best possible driver in mainline and not maintaining a separate version for older kernel versions, while we don't even know if somebody needs that at all.

People installing Linux nowadays on their Macs will probably use a kernel >=4.14 anyway and people already running Linux <4.1.4 can always refer to an older version of the driver as mentioned in the README. And of course they only need to do that if they want to update the driver.

jamesalmeida:
I wholeheartedly agree with @Dunedan

Dunedan:
As @roadrunner2's driver was recently merged upstream this PR isn't of much use anymore.

